PR #4, opened 2020-05-23
Add _byehook macro

robertbachmann:
- Before '\bye' print a warning if the user forgot to load a Unicode font.
- Before '\bye' print a warning if there were unresolved references.

### DIFF
--- a/optex/base/basic-macros.opm
+++ b/optex/base/basic-macros.opm
@@ -97,10 +97,14 @@
    if there were unresolved references.
    \_cod ------------------------------
 
-\_def\_byehook{%
+\_def\_warnunifont{%
 \_ifx\_initunifonts\_relax \_relax\_else%
 \_opwarning{You forgot to load a Unicode font}%
 \_fi%
+}
+
+\_def\_byehook{%
+\_warnunifont%
 \_ifnum\_unresolvedrefs>0%
 \_opwarning{Rerun to get references right}%
 \_fi%
--- a/optex/base/basic-macros.opm
+++ b/optex/base/basic-macros.opm
@@ -92,6 +92,20 @@
 
 \_public \loggingall \tracingall ; 
 
+   \_doc ------------------------------
+   Write a warning if the user forgot to load a Unicode Font {\em or} 
+   if there were unresolved references.
+   \_cod ------------------------------
+
+\_def\_byehook{%
+\_ifx\_initunifonts\_relax \_relax\_else%
+\_opwarning{You forgot to load a Unicode font}%
+\_fi%
+\_ifnum\_unresolvedrefs>0%
+\_opwarning{Rerun to get references right}%
+\_fi%
+}
+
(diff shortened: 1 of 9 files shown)


optex/base/basic-macros.opm
```diff
@@ -92,6 +92,20 @@
...
 \_public \loggingall \tracingall ; 
 
+   \_doc ------------------------------
+   Write a warning if the user forgot to load a Unicode Font {\em or} 
+   if there were unresolved references.
+   \_cod ------------------------------
+
+\_def\_byehook{%
+\_ifx\_initunifonts\_relax \_relax\_else%
+\_opwarning{You forgot to load a Unicode font}%
```
olsak, line 102: Maybe user want to leave only default 8bit fonts. For example, there is only testing run. It means, that user don't forgot. If user really forget to load Unicode font and use unicode characyers then there are a lot of missing character messages (not only in the log bat in the terinal too). Do you decide that this message is useful in some cases?

robertbachmann: When the user uses a char like 'ž' then there will be warning:
`    Missing character: There is no ž (U+017E) in font ec-lmr10!`
There are certain chars like '©' or 'ß' that will _not_ produce a warning, but the chars will be mapped incorrectly

````tex
%!TEX program=optex
%\fontfam[lmfonts]
\begitems
* {\Red £ © ¥ § ¶}
* 20°C {\Red °}
* {\Red «text» }
* German ä ö ü {\Red ß }
* Spanish: á, é, í, ó, ú, ü, ñ, {\Red ¿, and ¡ }
* Dutch: Ë/ë, {\Red ÿ}
* Porto.: {\Red ª, º}
\enditems
\bye
````
However its possible that some users (still) want to use the 8bit fonts ...
I have added a commit that extracts a ` \_warnunifont` macro, so
that such users could use

````tex
\let\_warnunifont\relax
````

olsak: Yes I understand. But I suggest to not implement new control sequence \_warunifont, only to re-folmulate the warning: from "You forgot to load a Unicode font" to "Unicode font was not loaded". An I am not friend with the % at the end of lines in the code if they need not to be here, like \fi%, \openref% etc. Maybe Ill correct this. Finally, I am not sure that all types of unresolved references were taken into account.

robertbachmann: I've changed the warning message. 

I've searched the code for 'TeX me again' and added the \_incr at 3 locations

Regarding the trailing '%', I've left them for now... because I was not a 100% sure where I need them and where not.

olsak: Bad behaviour of ß in preloaded is serious problem. I thought about 7 bit fonts preloaded (missing character will be here) but CM fonts are not good alternative because they are not ASCII encoded which seems impractical for testing. For example for printing \meaning\macro.

PR #13, opened 2020-12-12
Fixed typos

karol-bujacek:
\fixedrm is used only twice. It is probably just a typo and it should be \rmfixed.


olsak:
Thank you very much. I included this to my larger commit because I have strarted more corrections and improvements of Font selection system and its documentation on my local computer already. I discovered independently the same issue about \fixedrm->\rmfixed.

PR #23, opened 2021-02-10
Patch 3

cao-:
I think this was your intention when you added `\incr` and `\decr`.  If it wasn't, then discard my pull request.


olsak:
This is good idea, thanks. I'll use it. I chceked it inconsistently when \decr was introduced.

PR #35, opened 2021-03-18
Allow text-mode \{ and \}

robertbachmann:


olsak:
Thank you for idea. I implemented it slightly different.

PR #40, opened 2021-04-06
Typos corrected

cao-:
While reading the first part of the technical documentation I took note of the spelling mistakes that I found.  I also tried to correct some English, but beware that I am not a native English speaker, so make your own judgement about them.  Let me know if this is useful.   In case it is, when I read the rest of the documentation I'll do the same.  I could also keep the spelling corrections and English "corrections" separated in different commits if it is better.


olsak:
Thank you. It is very useful for me. My English is poor. I don't see important difference between your two suggestions, English corrections can be mixed with typos correction in single commits.

PR #45, opened 2021-04-24
Add sanity check to namespace export macros.

robertbachmann:
`\public \foo` would make `\let\foo=\relax` in cases were no
corresponding `\_foo` was defined.
This would hide typing errors.

* Change `\_public \foo ;` so that we get an error message if there is
no `\_foo`.
Users that want the old semantic can explicitly write `\let\foo=\relax`
* Change `\_nspublic \foo ;` so that we get an error message if there is
no `\.foo`.
* Change `\_private \foo ;` and `\_nsprivate \foo ;` so that we get an
error message if there is no `\foo`.

### DIFF
--- a/optex/base/math-macros.opm
+++ b/optex/base/math-macros.opm
@@ -218,7 +218,7 @@
 \_mathchardef\smallint="1273
 
 \_mathchardef\triangleleft="212F   \_private \triangleleft;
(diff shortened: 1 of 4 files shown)


olsak:
Thank you for this idea. Normally, OpTeX macros generates only warnings but this should be an exception because `\private` etc. are not used in normal document processing.

PR #60, opened 2021-07-09
Manage graphics states

vlasakm:
PDF page resources (like ExtGStates) in the pdfTeX model are settable only using `\pdfpageresources` primtive token register. While that is enough if the register has a single user, it requires a more thorough mechanism for anything more substantial.

I think in OpTeX we reached that point. Shadows and slide layers already used the register, now with the (coming) addition of transparency OpTeX trick and already existing clash with TikZ we have to do something about it.

The implemented solution either hooks into PGF/TikZ's way of managing PDF page resources or creates our own (very similar) mechanism. The principle is to set `\pdfpageresources` to point to pretty much a single global dictionary of ExtGStates (or other resources) that applies to all pages.

The global idea is of course prone to naming clashes. Another catch is that, for example only one value of shadow transparency can be used per document. But that was the case previously as well.

### DIFF
--- a/optex/base/graphics.opm
+++ b/optex/base/graphics.opm
@@ -319,6 +319,39 @@
 \_public \inoval \incircle \ratio \lwidth \fcolor \lcolor \shadow \overlapmargins ;
 
    \_doc -----------------------------
+   Just before defining shadows, which require special graphics states, we
+   define means for managing these graphics states. This is important, because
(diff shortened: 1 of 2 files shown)


olsak:
Thanks, it seems more better than original implementation of features based on `\pdfpageresources`.

PR #63, opened 2021-07-15
Add basic TikZ compatibility layer

vlasakm:
This is notably incomplete, because a similar `pgf.opm` would be needed
for completeness. This should however suffice for regular usage.


vlasakm:
This can be used as a test:

```
\fontfam[lm]
\load[tikz]

% From PGF manual, section 17.13.2 "Referencing the Current Page Node - Absolute Positioning"
\tikzpicture [remember picture,overlay]
\draw [line width=1mm,opacity=.25] (current page.center) circle (3cm);
\endtikzpicture

\bye
```

olsak:
I used your code, thank you.

PR #68, opened 2021-08-01
Retire the old microtypography trick

vlasakm:
Some details are mostly obsolote for LuaTeX with OpenType fonts, to qoute
the LuaTeX manual:

> In fact, the primitives that create expanded or protruding copies are
> probably only useful when used with traditional fonts because all
> these extra OpenType properties are kept out of the picture.

### DIFF
--- a/web/optex-tricks.html
+++ b/web/optex-tricks.html
@@ -212,28 +212,10 @@ The important OPmac tricks will be re-implemented here soon.
 <h2><a name="fontexpand"></a>Micro-typography: font expanding, hanging punctuation</h2>
 
 <p>LuaTeX supports "microtypography features": slight font expanding when
-the paragraph line is stretched and hanging punctuation. You can enable this
-using primitives \pdffontexpand+\adjustspacing and
-\rpcode,\lpcode+\protrudechars. Example:
-
-<pre>
-\pdffontexpand\_tenrm 30 20 5   % font expanding configured for \_tenrm
-\adjustspacing=2                % font expandnig activated
-
-\chardef\hyph=\hyphenchar\_tenrm
-\rpcode\_tenrm\hyph=310  \rpcode\_tenrm`\.=200 \rpcode\_tenrm`\,=200
-\lpcode\_tenrm`\“=400 \rpcode\_tenrm`\”=400  % hanging punctuation configured for \_tenrm
-\protrudechars=2                             % hanging puctuation activated
-</pre>
-
-<p>See <a href="http://petr.olsak.net/ftp/olsak/optex/tex-nutshell.pdf">TeX in a Nutshell,
-page 23</a> for more information about parameters of used primitives.
-
-<p>You can start from this example and modify it for your needs.
-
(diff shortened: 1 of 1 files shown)


olsak:
The want to keep the elementary example about micro-typographical features. The mte.opm package is linked already.

vlasakm:
I checked again. My understanding was flawed. LuaTeX in fact doesn't copy fonts to achieve font expansion. Hence `\pdffontexpand` is not wrong even for OpenType fonts, it works just by editing the Lua font table, which is what the `mte` package also does.

But I discovered another interesting piece of information from [evenmore.pdf, chapter 6](https://www.pragma-ade.com/general/manuals/evenmore.pdf):

> There is no need to have a different step than 1. In pdfTeX instances are created per step used, but in LuaTeX this is more fluid. There is no gain in using different steps. We just keep it for compatibility.

I think that the example here and the `mte` package should be adjusted.

Same document also mentions:

> There is one drawback with this method, although so far I never heard a user complain, which can be an indication of how this mechanism is used: you cannot mix fonts with different step, stretch and/or shrink. As we just did this in the example, this statement is not really true in LuaMetaTEX: there we only need to keep the step the same.

My understanding is probably wrong again, because something like the following seems to work out OK:

```
\def\SetLipsumLanguage#1{}
\fontfam[lm]
\input nodetree
\NodetreeRegisterCallback{preout}
%\it\bf\rm

\expandglyphsinfont\_tenrm 30 20 1
\expandglyphsinfont\_tenbf 30 20 1
\adjustspacing=2

test {\bf this} {\em text} \lipsum[1]

\meaning\_tenrm

\meaning\_tenbf

\meaning\_tenit

\bye
```

output for the first line:

```
/F37 9.96264 Tf
0.989 0 0 1 90.791 761.061 Tm [<0069003200620069>]TJ
/F45 9.96264 Tf
1 0 0 1 110.021 761.061 Tm [<0069003F00420062>]TJ
/F46 9.96264 Tf
1 0 0 1 131.848 761.061 Tm [<0069003200740069>]TJ
/F37 9.96264 Tf
0.989 0 0 1 151.584 761.061 Tm [<0047005100600032004B>-335<004200540062006D004B>-336<002F0051004800510060>-335<006200420069>-336<001C004B00320069002D>-336<002B0051004D00620032002B006900320069006D00320060>-335<001C002F0042005400420062002B0042004D003B>-336<00320048004200690058>-336<006C0069>-335<0054006D0060006D0062>-336<0032004800420069002D>-335<0070>28<00320062006900420023006D0048006D004B>-335<006D0069002D>]TJ
```

Notice, that I didn't expand `\_tenit`.

Also notice, that `\_tenbf` has not in fact been expanded either! It's because `\_tenrm` at the time of `\expandglyphsinfont` is actually the preloaded font... Very subtle, I found this out only by chance (\nullfont can't be expanded and there are no preloaded fonts in `mmoptex`).

I hope this shows that the example in this trick is too simplistic to be useful. If you decide to not delete it, I suggest at least changing `\pdffontexpand` to `\expandglyphsinfont`.

PR #69, opened 2021-08-01
Fix for new lipsum version

vlasakm:
New lipsum version starts the file with `\SetLipsumLanguage{latin}`.


olsak:
Thank you for noticing about it.

PR #78, opened 2021-09-23
fix section sign (\S) Unicode code point

balejk:


olsak:
Thank you for noticing about this problem. I solved it in my last commit https://github.com/olsak/OpTeX/commit/62a36aba58136c78c925d9f5ea7434e636a45422

balejk:
Thank you for fixing this and please forgive my nonsensical fix, which would break behavior without Unicode fonts.

PR #88, opened 2022-01-28
Add access to callback internals

vlasakm:
This may be used for integration with other packages, like minim.


olsak:
Why there is `local saved_luatexbase`? I don't se any usage of the identifier `saved_luatexbase`
.

vlasakm:
> Why there is `local saved_luatexbase`? I don't see any usage of the identifier `saved_luatexbase`

Oops, sorry you are right. That was a remnant of a different approach I tried.

PR #89, opened 2022-02-04
A couple of small fixes

vlasakm:
A few typos + corrections in the direction of only logging, not printing on terminal.


olsak:
Done.

PR #97, opened 2022-03-02
Simplify Lua function allocation

vlasakm:
This changes to the classic appending to Lua table:

```
table[#table + 1] = something
```

### DIFF
--- a/optex/base/optex.lua
+++ b/optex/base/optex.lua
@@ -49,9 +49,8 @@ end
 -- Allocator for Lua functions ("pseudoprimitives"). It passes variadic
 -- arguments (\"`...`") like `"global"` to `token.set_lua`.
 local function_table = lua.get_functions_table()
-local luafnalloc = 0
 function define_lua_command(csname, fn, ...)
-    luafnalloc = luafnalloc + 1
+    local luafnalloc = #function_table + 1
     token.set_lua(csname, luafnalloc, ...) -- WARNING: needs LuaTeX 1.08 (2019) or newer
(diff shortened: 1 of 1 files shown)


olsak:
Where the initial value of luafnalloc is set? (if we remove the line `local luafnalloc = 0`).

vlasakm:
We set `function_table` to be a reference to table that LuaTeX has built in.

The details are more intricate, but the `#` operator essentially returns the biggest number `n` such that `table[1]`, ... `table[n]` are all not `nil` or `0`.

Because the table is empty initially, we get the initial zero and then we can add values with the classic Lua idiom:

```
table[#table+1] = value
```

So the sequence:

```
table[1] = value
table[2] = value
```

And we can get the number of elements in that table with `#table`. I think this is better because we don't keep any state that has to be potentially synchronised.

I am not sure if I remember it correctly, but I have seen something allocate a Lua function in a similar way I suggest above. If then somebody would allocate with OpTeX, and because the OpTeX counter wouldn't be updated, OpTeX would perform the allocation of the same register.

The approach above doesn't have the synchronization problem. But on the other side, it would be problematic if someone would start randomly deleting the Lua functions. Unlike LaTeX, OpTeX doesn't hand out the allocation numbers so this situation is even more unlikely.

olsak:
I done it in my last commit.

vlasakm:
Closed by https://github.com/olsak/OpTeX/commit/d0c577aabc2aca528574993b5f1018078c98a8c0

PR #103, opened 2022-12-21
Try to improve wording of OpTeX trick 99

vlasakm:
Try to make Lua function allocation more robust by introducing `luatexbase.new_luafunction` which is by interface compatible with the function of the same name from LaTeX (see `texdoc ltluatex`) . Though instead of using a TeX count register to keep the number of allocated Lua functions we use the natural length of the Lua functions table.

Directly returning the length of said table is not nice, since if the user doesn't add anything to the allocated entry, the length stays the same and we could return the same allocation number to some other caller. We can somehow mitigate this by setting the allocated entry to something -- a dummy function.

This can be used to simplify the pstricks OpTeX trick. While at it I tried to improve the wording.

I am unfortunately not anywhere near my current setup, so this pull request is untested. Hopefully I didn't miss anything fundamental and this can be merged after someone confirms that it works.

### DIFF
--- a/web/optex-tricks.html
+++ b/web/optex-tricks.html
@@ -1558,36 +1558,32 @@ QRcode.
 
 <h2><a name="pstricks"></a>Using pstricks</h2>
 
-<p>First of all: pstricks is obsolete method for programming pictures. Now,
-TikZ is used. OpTeX supports TikZ by \load[tikz]. So, the usage of pstricks
-mentioned here is only experimental.
+<p>First of all: pstricks is an obsolete method for programming pictures.
+Nowadays, TikZ has mostly replaced it. You can use TikZ in OpTeX simply with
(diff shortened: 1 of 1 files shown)


olsak:
I don't plan to add a declaration of luatexbase.new_luafunction to the optex lua base code. The OpTeX trick about pstricks is only very specific, I don't recommend to use it.
Thank you for improving text of OpTeX trick 0099, I'll use this commit.

vlasakm:
I rebased the changes and removed the changes with regards to `new_luafunction`.

olsak:
I don't plan to add a declaration of luatexbase.new_luafunction to the optex lua base code. The OpTeX trick about pstricks is only very specific, I don't recommend to use it.
Thank you for improving text of OpTeX trick 0099, I'll use this commit.

olsak:
done in another commit

PR #109, opened 2023-03-04
Inconsolata: font family file added

omasanori:
See also: https://www.ctan.org/pkg/inconsolata


omasanori:
The two variant should be merged like Roboto...

PR #110, opened 2023-03-04
Fira: font family file added

omasanori:
See also: https://ctan.org/pkg/fira and https://github.com/bBoxType/FiraSans


olsak:
Thank you very much!

Issue #108, opened 2023-03-03
Wrong \foreach while loading some libraries in tikz


cao-:
When trying to load some libraries in Tikz, using the  `\usetikzlibrary` command, the macro `\foreach` has the OpTeX definition rather than that of Tikz.  Here is an example that fails:
```tex
\load[tikz]
\usetikzlibrary{graphs}
\bye
```
It doesn't fail if i redefine `\foreach`:
```tex
\load[tikz]\let\foreach=\_pgf_foreach
\usetikzlibrary{graphs}
\bye
```

I see that in tikz.opm we have `\addto \tikz@startup@env {\let\foreach=\pgffor@foreach}`:  maybe there is some other Tikz macro that we should modify in the same way.

olsak:
Thank you for bug report. I tried to solve it by my last commit.

cao-:
Thank you for solving it :)

PR #111, opened 2023-03-04
Use proper \foreach while loading Tikz packages

cao-:
This could be a solution to #108 .  But I'm not sure this is the best way to deal with it.

### DIFF
--- a/optex/pkg/tikz.opm
+++ b/optex/pkg/tikz.opm
@@ -152,6 +152,11 @@
    \_cod
 
 \addto \tikz@startup@env {\let\foreach=\pgffor@foreach}
+\_def\pgfutil@InputIfFileExists#1#2#3{%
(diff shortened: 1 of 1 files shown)


cao-:
This PR is not relevant anymore and should be closed,  the issue was already solved with your last commit.  Anyhow, I take it as an opportunity to make you notice a potentially unwanted behaviour I spotted in the tikz.opm code.  I see that at the end we do `\_let\foreach=\_optexforeach`, so we enforce `\foreach` to have the default meaning defined in OpTeX.  Shouldn't we restore it to the meaning it has before loading this file?  Maybe the user already redefined it before doing `\load[tikz]` and doesn't want the default meaning to be enforced.

olsak:
Solved by commit https://github.com/olsak/OpTeX/commit/528991dee8cb52de6b7d477d2ce91a5d346c858c

PR #114, opened 2023-03-06
Sort non-GUST fonts by their names

omasanori:
The resulting diff is a mess, but the catalog becomes easy to find a typeface.


omasanori:
I think it would be better to expand the catalog with more fonts on TeX Live and reorganize it based on the foundries.

PR #120, opened 2023-03-08
Add free and open typefaces from indestructible type*

omasanori:
Website: https://indestructibletype.com/
GitHub: https://github.com/indestructible-type

- [x] [Besley*](https://github.com/indestructible-type/Besley) (slab serif)
- [x] [Bodoni*](https://github.com/indestructible-type/Bodoni) (Bodoni reproduction, supporting optical sizes)
- [x] [Drafting* Mono](https://github.com/indestructible-type/Drafting) (monospaced)
- [x] [Jost*](https://github.com/indestructible-type/Jost) (geometric sans-serif)

### DIFF
--- a/optex/base/f-besley.opm
+++ b/optex/base/f-besley.opm
@@ -1,0 +1,46 @@
+%% This is part of the OpTeX project, see http://petr.olsak.net/optex
+
+\_famdecl [Besley] \Besley {Besley* antique slab serif}
+        {\caps \allsc \nocaps \cond \narrow}
+        {\rm \bf \it \bi \mf \mi \ef \ei \kf \ki} {Schola}
+        {[Besley-Regular]}
+        {\_def\_fontnamegen {[Besley\_subV-\_currV]:script=latn;\_capsV\_fontfeatures}}
+
+\_wlog{\_detokenize{%
+Modifiers (width):^^J
+ \cond ....... Condensed variants^^J
+ \narrow ..... Narrow variants^^J
+Modifiers (small caps):^^J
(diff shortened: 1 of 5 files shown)


olsak:
These fonts are not in the TeXlive distribution. I suggest to create a command `\_famsrc {text}` which can be used in `fams-ini.opm` file in the similar sense as `\_famfrom`. The parameter `text` can include the URL or source specification, where the font can be found. When we generate the font catalog  then fonts are shown if they are installed. Else only the message is printed "the font family is available: text". The `\_famsrc {TeXlive}` can be set for almost all fonts but may be there can be different sources like in these fonts Besley, Drafting, and in the case of already provided Technika fonts. I plan to create such command.

olsak:
Besley font family provides variable font feature, see http://petr.olsak.net/optex/optex-tricks.html#varifonts for example. But I have bad experiences with variable fonts in general when `luatex` and `luaotfload` is used. For example, you can try to load  `Besley[wdth,wght].ttf`, rename it to `Besley-varfont.ttf` and do
```
\ufont\f=[Besley-varfont]:axis={wght=500,wdth=80}

\f This is test of variable font.

\bye
```
The result includes big bugs in rendering.  I don't know if this bug is in the `luaotfload` package and a new issue can be created in this project. In fact, LaTeX uses `luahbtex` instead `luatex` and uses variable fonts with HarfBuzz library.

omasanori:
Oh, I noticed just now that you had committed `\_famsrc`.

olsak:
Thank you.

PR #123, opened 2023-03-11
Add basic language data for Estonian

mattiasmarka:
I've added basic language data for Estonian, only the fields that made sense. Note that I've made \"text" and \'text' be equivalent in Estonian. I've explained why this is in the commit message. Alas, I couldn't figure out how to test these changes since (as I understand it) the language data is lazy-loaded from the TeX distribution.


olsak:
Thank you.

PR #125, opened 2023-03-14
Extend  `\_inkdefs`

robertbachmann:
# Commit: Extend  `\_inkdefs`
- add \textbf
- add \textit
- add \textsc
- add \transparent

These macros may be generated by Inkscape's `latex-text-renderer.cpp`.

# Additional notes

The relevant code can be seen here 
<https://gitlab.com/inkscape/inkscape/-/blob/master/src/extension/internal/latex-text-renderer.cpp#L390>

------------

Test case:

```tex
\fontfam[pagella]

\inkinspic{figure.pdf}

\bye
```

I'm using Inkscape 1.1.2 (0a00cf5339, 2022-02-04)
with 
``
inkscape figure.svg --export-type=pdf --export-latex -o figure.pdf      
``

Inkscapes generates 
````tex
  \begin{picture}(1,0.62019123)%
    \lineheight{1}%
    \setlength\tabcolsep{0pt}%
    \put(0,0){\includegraphics[width=\unitlength,page=1]{figure.pdf}}%
    \put(0.80999305,0.15562135){\color[rgb]{1,0.4,0}\makebox(0,0)[lt]{\lineheight{1.25}\smash{\begin{tabular}[t]{l}Some\\\textbf{\textit{tex}}\textbf{\textit{t}}\end{tabular}}}}%
    \put(0.08211842,0.48438098){\color[rgb]{0,0,0}\makebox(0,0)[lt]{\lineheight{1.25}\smash{\begin{tabular}[t]{l}regular, \textbf{bold},\\\textit{italic}, \textbf{\textit{bold-italic }}\\\end{tabular}}}}%
    \put(0.05959094,0.13101){\color[rgb]{0,0.50196078,0}\makebox(0,0)[lt]{\lineheight{1.25}\smash{\begin{tabular}[t]{l}\textsl{slanted, }\textbf{bold}\\\textbf{\textsl{bold slanted}}\end{tabular}}}}%
    \put(0.60685555,0.51537681){\color[rgb]{0,0,0}\transparent{0.40999392}\makebox(0,0)[lt]{\lineheight{1.25}\smash{\begin{tabular}[t]{l}semi-transparent\end{tabular}}}}%
  \end{picture}%
````

I've attached the SVG file.

[figure.zip](https://github.com/olsak/OpTeX/files/10974337/figure.zip)

### DIFF
--- a/optex/base/graphics.opm
+++ b/optex/base/graphics.opm
@@ -59,6 +59,11 @@
   \_def\rotatebox#1#2{\_pdfrotate{#1}#2}%
   \_def\lineheight#1{}%
   \_def\setlength#1#2{}%
+  \_def\transparent#1{}%
+  % Inkscape may generate \textbf{\textit{\textsc{TEXT}}}
+  \_def\textbf#1{\_begingroup\_let\_it\_bi\_bf #1\_endgroup}%
+  \_def\textit#1{\_begingroup\_it #1\_endgroup}%
+  \_def\textsl#1{\_begingroup\_trycs{slant}{}\_it #1\_endgroup}%
 }
 \_public \inkinspic ;
 
--- a/optex/base/graphics.opm
+++ b/optex/base/graphics.opm
@@ -59,7 +59,7 @@
   \_def\rotatebox#1#2{\_pdfrotate{#1}#2}%
   \_def\lineheight#1{}%
   \_def\setlength#1#2{}%
-  \_def\transparent#1{}%
+  \_def\transparent#1{\_ea\_transparency\_expanded{\_expr[0]{(1-.5)*255}} }%
   % Inkscape may generate \textbf{\textit{\textsc{TEXT}}}
   \_def\textbf#1{\_begingroup\_let\_it\_bi\_bf #1\_endgroup}%
   \_def\textit#1{\_begingroup\_it #1\_endgroup}%

### CODE REVIEW

File: optex/base/graphics.opm [hunk 1]
```diff
@@ -59,6 +59,11 @@
   \_def\rotatebox#1#2{\_pdfrotate{#1}#2}%
   \_def\lineheight#1{}%
   \_def\setlength#1#2{}%
+  \_def\transparent#1{}%
```
olsak, line 62: I don't know what `\tranasparent` does mean in LaTeX, maybe: `\def\transparent#1{{\_transparency255 #1}}` ?

robertbachmann:
The `\transparent` macro used by the inkscape generated file is from `transparent.sty`. http://mirrors.ctan.org/macros/latex/contrib/transparent/transparent.pdf 

I takes a number from 0 -- 1, but careful 1 means 100% opaque:

> 1. User interface
> The package transparent defines \transparent and \texttransparent. They are
> used like \color and \textcolor. The first argument is the transparency value
> between 0 and 1 where 0 is fully transparent and 1 is opaque.

I implemented it like this.

```tex
\_def\transparent#1{\_edef\_tmpt{\_expr[0]{(1-#1)*255}}\_transparency=\_tmpt}%
```

Note sure if there's an easy way to do it without the `\_tmpt`

-----------

Test case:

```tex
\fontfam[lm]
\inkinspic{figure.pdf}
\bye
```

LaTeX test file I used:
```tex
\documentclass{article}
\usepackage{graphicx}
\usepackage{xcolor}
\usepackage{transparent}
\begin{document}
\input{figure2.pdf_tex}
\end{document}
```

Second SVG file
[figure2.zip](https://github.com/olsak/OpTeX/files/10984442/figure2.zip)

Sceenshot of result: 
image

olsak:
You can use this code
```
\_def\transparent#1{\_ea\_transparency\_expanded{\_expr[0]{(1-.5)*255}} }
```
This is without `\_tmpt`. If the `\expr` macro was expandable in classical sense then simply `\transparency=\expr...` would be enough. But it is not true in this case because `\immediateassigment` is used in `\expr` and this LuaTeX primitive is (wrongly) applied to the currently executed assignment `\transparency=` not to the desired assignment inside the `\expr` macro. I didn't know about this "feature" yet and I am not sure if it is feature or bug in LuaTeX.

olsak:
Or we can use
```
\_def\transparent#1{\_transparency\_exprA[0]{(1-.5)*255} }
```
but we cannot be sure if the `\_exprA` "API" will be stable in OpTeX. Maybe yes.

olsak:
I did a more thorough investigation and I found that the issue is due to `\afterassignment` used in the `\trnasparency` macro. This primitive is applied to the first assignment done inside `\expr ` macro and not to the `\_transpattr=` assignment. Now, I am sure this is feature, not bug.

robertbachmann:
Thanks for the explanation.

PR #126, opened 2023-03-16
Fix tests

omasanori:
Newly landed tests failed on my machine, for an understandable reason.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+optex/build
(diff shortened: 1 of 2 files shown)


Bob-AT:
Good catch. 

I tested on my system, but forgot that I had an OpTeX installed in `/usr/local/texlive/2022`, so on my system it worked because kpse was resolving `op-ring.png` to  `/usr/local/texmf-dist/tex/optex/demo/op-ring.png`

robertbachmann:
Good catch. 

I tested on my system, but forgot that I had an OpTeX installed in `/usr/local/texlive/2022`, so on my system it worked because kpse was resolving `op-ring.png` to  `/usr/local/texmf-dist/tex/optex/demo/op-ring.png`.

PR #138, opened 2023-06-29
Clara serif font family added.

petrk23:
Clara is a font family created specially by Séamas Ó Brógáin
for printing the book "A Dictionary of Editing" in 2015.
The font is a little heavier, but it's good usable and readable.
It also supports the Greek and Cyrillic alphabets, so it can be
called multilingual. The glyphs have correct accents and kerning.
Czech, for example.

The best matching math font is XITSMath. The only difference
is in few glyphs, but the others are the same or very similar.
It has a little less x-height, but only a little.

Maybe a better description would be good!

It is available from CTAN:
https://ctan.org/pkg/clara

### DIFF
--- a/optex/base/f-clara.opm
+++ b/optex/base/f-clara.opm
@@ -7,7 +7,7 @@
 
 \_wlog{\_detokenize{%
 Modifiers:^^J
- \caps ...... caps & small caps (\rm, simple alphabet only!)^^J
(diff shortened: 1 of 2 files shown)


petrk23:
The last commit with a newline was added to match most other font files. The question is: it is necessary to have these double newlines at the end?

olsak:
Thank you.

PR #145, opened 2023-08-01
OMLS Lua implementation

omasanori:
*No real thing is happening yet, stay tuned.*

I am starting to write a Lua OMLS parser. It will support standalone and [Pandoc filter](https://pandoc.org/lua-filters.html) usages. I assume that, if accepted, @olsak uploads ZIP files on petr.olsak.net like OpTeX for archiving and [submits .rockspec files on LuaRocks](https://github.com/luarocks/luarocks/wiki/Creating-a-rock#submitting-a-rockspec-for-inclusion-in-the-rocks-server) so that one can install the parser by executing `luarocks install omls`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 optex/build
+lua/omls-*
--- a/lua/Makefile
+++ b/lua/Makefile
@@ -1,0 +1,23 @@
+.POSIX:
+.SUFFIXES:
+
+NAME = omls
+VERSION = 0.1.0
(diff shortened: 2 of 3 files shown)


olsak:
It will be great. I'll test the result and add the link to the sources at my web page. But the source files themselves should be published in such a way that they are under the control of the author.

omasanori:
@olsak I see. Then, I will make a dedicated repository. Thank you for your suggestion!

PR #154, opened 2023-11-02
Patch `\newcatcodetable` before loading minim-alloc

vlasakm:
In the latest release minim actually uses `\newcatcodetable` in minim-alloc.tex. This means that it still uses the standard OpTeX definition (which doesn't initialize the catcode table), but it expects the initialization.

We solve this by doing the patch to `\newcatcodetable` _before_ loading minim-alloc.tex. It should have probably been done this way right from the beginning, it just didn't manifest.

### DIFF
--- a/optex/pkg/minim.opm
+++ b/optex/pkg/minim.opm
@@ -93,6 +93,18 @@
 }
 
    \_doc
+   Both \LaTeX/ and the minim inspired catcode table allocators initialize the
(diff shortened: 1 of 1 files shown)


vlasakm:
The "problematic" minim version was released two weeks ago, so this hasn't been problem for that long.

I found it while preparing a new release of [mmtex](https://github.com/vlasakm/mmtex/) with my `optex-minim` test file: https://github.com/vlasakm/optex-minim/blob/master/examples/optex-minim-pgf-test.tex (in case you want to test yourself).

Issue #173, opened 2024-04-30
A bug in `\replstring`


Udi-Fogiel:
When we try to do `\replstring\foo{<find>}{<replace>}`, and `\foo` has a sequence of the form `<find>{<braced content>}<find>`, then `\replstring` strips the braces off. For example
```tex
\def\foo{ab{c}b}
\replstring\foo{b}{d}
\show\foo
\bye

> \foo=macro:
->adcd.
```

It can also happens in other situation, as long as the braced content is passed into a macro as the only argument it will happen. Another example is 
```tex
\def\foo{{a}b}
\replstring\foo{b}{d}
\show\foo
\bye

> \foo=macro:
->ad.
```
It is a problem in situation where the braces are important, for example in [OpTeX trick 38](http://petr.olsak.net/optex/optex-tricks.html#circletext) the equations must be inside groups. 

If we will modify slightly the example in the description of the trick, it will break. For example if we would like to add a space after a bullet
```tex
\hbox{%
\circletext {1.7cm} {212}  {{$\bullet$} UNIVERSITAS  CAROLINA PRAGENSIS {$\bullet$} }
                           {\spaceskip=.5em \kpcirc TA{-.1}\kpcirc NA{-.05}}
\circletext {-1.7cm} {237} {Facultas MFF}
                           {\kpcirc Fa{-.1}}
}
``` 
or add another bullet in the middle
```tex
\hbox{%
\circletext {1.7cm} {212}  {{$\bullet$} UNIVERSITAS {$\bullet$} CAROLINA PRAGENSIS {$\bullet$}}
                           {\spaceskip=.5em \kpcirc TA{-.1}\kpcirc NA{-.05}}
\circletext {-1.7cm} {237} {Facultas MFF}
                           {\kpcirc Fa{-.1}}
}
```

I suggest a new approach to implement `replstring`:
```lua
\directlua{%
local scan_toks = token.scan_toks
local put_next = token.put_next
local create = token.create
local lbrace = create(string.byte('{'))
local rbrace = create(string.byte('}'))
define_lua_command("replstring", function()
    local macro_name = create(token.scan_csname())
    put_next(create('_expandafter'),
    lbrace,macro_name,rbrace)
    local macro_body = scan_toks()
    local nested = token.scan_keyword('nested')
    local find = scan_toks()
    local replace = scan_toks()
    local find_length = \csstring\#find
    local replace_length = \csstring\#replace
    local range = \csstring\#macro_body - find_length + 1
[...]

PR #174, opened 2024-05-01
reimplement \replstring in lua

Udi-Fogiel:
See the discussion in #173

### DIFF
--- a/optex/base/more-macros.opm
+++ b/optex/base/more-macros.opm
@@ -377,6 +377,7 @@
 
 \_endcode % -------------------------------------
 
+2024-05-01 \replstring reimplemented in lua
 2024-02-10 \optdef reimplemented, it can create expandable macros now.
 2024-01-18 \bracedparam reimplemented.
 2023-12-04 \nnum introduced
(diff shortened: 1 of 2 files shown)


optex/base/more-macros.opm
```diff
@@ -194,42 +194,17 @@
...
-   \_long\_def\_replacestringsA##1#2{\_tmptoks{##1}\_replacestringsB}%
-   \_long\_def\_replacestringsB##1#2{\_ifx!##1\_relax \_else \_toksapp\_tmptoks{#3##1}%
-                                     \_ea\_replacestringsB\_fi}%
-   \_ea\_replacestringsA #1?#2!#2%
-   \_long\_def\_replacestringsA##1?{\_tmptoks{##1}\_edef#1{\_the\_tmptoks}}%
-   \_ea\_replacestringsA \_the\_tmptoks}
-\_normalcatcodes
-
-\_public \replstring ;
+\_def\replstring{\_replstring}
```
Udi-Fogiel, line 207: @olsak I wasn't sure if using `\_let` would be better here, because then the order of the files would be significant (and I'm not sure how lua code interacts at format generation time as well), so I just went with `\_def` as I saw it was done with `\_addpageresource`.

PR #181, opened 2024-06-02
Allow multiple callbacks with the same description

vlasakm:
This cautious check for multiple entries was added in last commit, and it turned out to be problematic with our `\mathsbon` callback, which can be activated multiple times in nested `\load`s (see #180).

While it seems better to me, to be proactive about the checks, LuaLaTeX also seems to allow duplicate entries, and results in same behavior when removal is involved (the first registered is also the first to be removed). So, let's just remove the check.

### DIFF
--- a/optex/base/optex.lua
+++ b/optex/base/optex.lua
@@ -235,12 +235,6 @@ function callback.add_to_callback(name, fn, description)
         err("missing description when adding a callback to '%s'", name)
     end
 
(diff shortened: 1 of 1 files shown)


vlasakm:
Closing in favor of https://github.com/olsak/OpTeX/commit/86bbcfd3e7f6aaab3a5029668b92df852ff2224a.

PR #191, opened 2024-09-10
Plex font family support added

petrk23:
Plex is a font family created as the corporate typeface of IBM. Fortunately, it was released under an OFL license, so we can use it too. The font is simply available from CTAN.

It's a high-quality font family (serif, sans, mono) with many weight variants ranging from thin to bold. During my work with it, I haven't spotted any problems with these fonts. The only thing missing are small caps.

The font is different, especially the serif and its italics, from the other fonts we already have. This is a clear argument for adding it to OpTeX, but the disadvantage is that there is no visually compatible math font available.

Under the line, we get another high-quality font family for OpTeX.

### DIFF
--- a/optex/base/f-plex.opm
+++ b/optex/base/f-plex.opm
@@ -1,0 +1,56 @@
+%% This is part of the OpTeX project, see http://petr.olsak.net/optex
+
+\_famdecl [Plex] \Plex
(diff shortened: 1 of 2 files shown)


olsak:
Thank you.

PR #200, opened 2025-01-11
Don't return true in list callbacks

Udi-Fogiel:
I guess whan the code was written
this was compatible with LaTeX's code,
but LaTeX has changed since then:
https://github.com/latex3/latex2e/commit/c038181414defe3b335380cc9521f2c692bd8012 
and handeling the 'true' case is kind of annoying
from a user perspective

### DIFF
--- a/optex/base/optex.lua
+++ b/optex/base/optex.lua
@@ -415,17 +415,15 @@ function callback.call_callback(name, ...)
 
     local head = (...)
     local new_head
-    local changed = false
     for _, fn in iter(functions) do
         new_head = fn(head, select(2, ...))
         if new_head == false then
             return false
         elseif new_head ~= true then
             head = new_head
-            changed = true
         end
     end
-    return not changed or head
+    return head
 end
 call_callback = callback.call_callback
 --


olsak:
@vlasakm Is it OK from your point of view?

vlasakm (approved):
Looks good to me!

Any "synchonisation" with LaTeX is I think desirable. The point of the callback mechanism in OpTeX is not only to provide users with ability to use callbacks, but to do it "similarly enough" to LaTeX such that things like `luaotfload` work.

I know there have been even more changes in upstream LaTeX about the callbacks, but unfortunately I haven't been really keeping OpTeX up :(

Udi-Fogiel:
I have also slightly simplified `mlist_to_hlist`.

olsak:
Sorry, I am not ready to merge this to main unless I understand what the code touched by second commit means. The original code looks like this
```
    local new_head = call_callback("pre_mlist_to_hlist_filter", head, ...)
    if new_head == false then
        node.flush_list(head)
        return nil
    elseif new_head ~= true then
        head = new_head
    end
```

It seems curious to me: If we have two-value logic then `new_head == false` is equivalent to `new_head ~= true`, so the elseif part is never executed. I am missing something? The similar code is at lines 427 and 459. The patch suggested by Udi changes behavior of the code, IMHO.

Udi-Fogiel:
Before 588c4b3 there were cases where `new_head` was `ture`, so the old logic had three cases:

* If `new_head` is `false` exit early.
* if `new_head` is a node (the case where `new_head` is not `false` and not `true`) then update `head`
* if `new_head` is `true` keep head as it is
After 588c4b3 the last case does not happen, as list callbacks will not return true, so if `new_head` is not false you can always return the node returned from the list callback.

PR #204, opened 2025-02-21
Fix coloring of outlined fonts

vlasakm:
We try to be lazy and only set emit color changing PDF literals when we encounter some TeX node that actually needs the color setting.

For glyph nodes we always returned that fill color is needed and that stroke color is not needed.

This is not correct though, as depending on what `mode` [1] the font is using, it may be filled or stroked. The mode of fill+stroke is used for example by the luaotfload embolden feature [2].

Although it's a bit expensive to check the mode for all the glyphs, it's the simplest way to get correct coloring in presence of outlined fonts implemented by LuaTeX.

[1]: LuaTeX manual, section "The font tables"
[2]: https://github.com/latex3/luaotfload/blob/72561365e3de47c138418af7d11ab335683ff7f0/src/luaotfload-embolden.lua?plain=1#L18

### DIFF
--- a/optex/base/optex.lua
+++ b/optex/base/optex.lua
@@ -39,6 +39,7 @@ local getid = direct.getid
 local getlist = direct.getlist
 local setlist = direct.setlist
 local getleader = direct.getleader
+local getfont = direct.getfont
 local getattribute = direct.get_attribute
 local setattribute = direct.set_attribute
 local insertbefore = direct.insert_before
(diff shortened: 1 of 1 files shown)


Udi-Fogiel:
I think this means OpTeX trick 0064 now works with colors. Should the line "Sadly outline fonts don't work with colors." be removed from the web page?

PR #211, opened 2025-11-25
Add pre_append_to_vlist_filter callbck

Udi-Fogiel:
This is similar to `pre_mlist_to_hlist`, useful if several users wants to use `append_to_vlist_filter` simultaneously, as it is exclusive.

### DIFF
--- a/optex/base/optex.lua
+++ b/optex/base/optex.lua
@@ -17,6 +17,8 @@ local fmt = string.format
 local node_id = node.id
 local node_subtype = node.subtype
 local flush_node = node.flush_node
+local flush_list = node.flush_list
+local node_traverse = node.traverse
 local glyph_id = node_id("glyph")
 local rule_id = node_id("rule")
 local glue_id = node_id("glue")
@@ -91,10 +93,10 @@ end
 -- It is used in the \^`\_rawht` macro.
 
 function optex.raw_ht()
-    local nod = tex.box[token.scan_int()]       -- read head node of the given box
+    local nod = tex.box[token_scanint()]       -- read head node of the given box
     local ht = 0
-    if nod ~= nil and nod.id == vlist_id then
-        for n in node.traverse(nod.head) do
+    if nod and nod.id == vlist_id then
+        for n in node_traverse(nod.head) do
             if n.id == hlist_id or n.id == vlist_id or n.id == rule_id then
                 ht = ht + n.height + n.depth
             elseif n.id == glue_id then
@@ -446,7 +448,7 @@ callback_register("mlist_to_hlist", function(head, ...)
     -- pre_mlist_to_hlist_filter
     local new_head = call_callback("pre_mlist_to_hlist_filter", head, ...)
     if new_head == false then
-        node.flush_list(head)
+        flush_list(head)
         return nil
     else
         head = new_head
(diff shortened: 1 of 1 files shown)


web/optex-tricks.html
```diff
@@ -926,6 +927,66 @@ <h2><a name="glyphmove"></a>Move glyphs by any offset</h2>
...
+you can do the following:
+
+<pre>
+\directlua{
+local funcs = lua.get_functions_table()
+local luafunction_count = \csstring#funcs
+luatexbase.new_luafunction = function(name)
+   luafunction_count = math.max(\csstring#funcs,luafunction_count) + 1
+   return luafunction_count
+end
```
vlasakm: Idea: Should this be just included in `optex.lua`? There are probably more packages that depend on this IMO unfortunate `new_luafunction` interface.

Question: My memory is blurry, would this work with what we do for minim package?

web/optex-tricks.html
```diff
@@ -926,6 +927,66 @@ <h2><a name="glyphmove"></a>Move glyphs by any offset</h2>
...
+luatexbase.callbacktypes.pre_append_to_vlist_filter = "list"
```
vlasakm: Question: Maybe I am missing something, but `pre_append_to_vlist_filter`'s type is already set by:

```lua
callback.create_callback("pre_append_to_vlist_filter","list")
```

no?

Udi-Fogiel: Yes, this line does not define the type (as luatexbase.callbacktypes.pre_append_to_vlist_filter is just a dummy variable), it is just that lua-ul checks for the existence of this callback this way:
https://github.com/zauguin/luaul/blob/cd887430ba53f91a3a7682c9cf8a0f8437dc708a/lua-ul.dtx#L235-L237

web/optex-tricks.html
Udi-Fogiel: you mean this?
https://github.com/olsak/OpTeX/blob/e964f135cb2b68fd9f9da6460f16ebc7293e2c22/optex/pkg/minim.opm#L89-L93

pretty similar to LaTeX's luatexbase.new_luafunction, but I think LaTeX don't assume that the count and the table size is synchronized (hence the math.max):
https://github.com/latex3/latex2e/blob/bb337366429f15dc40ff39b70d9febed0f1930ad/base/ltluatex.dtx#L1304-L1308

So this will probably work for both.

Udi-Fogiel:
I think it is finalized.

olsak:
Now, I have message: This branch cannot be rebased due to conflicts. Maybe, it is due to I accepted the issue "use local variables to avoid...". Can you please take a look at this?

Udi-Fogiel:
I have resolved the conflict in this PR, did you pull the latest commit ([Merge branch 'master' into pre_append_to_vlist_filter](https://github.com/olsak/OpTeX/pull/211/commits/373bc0c5a96440b95d229f8354ed52da963a7271))? I'll take this opportunity to fix a couple of typos :)

Udi-Fogiel:
> I'll take this opportunity to fix a couple of typos :)

never mind, It will be easier to do that after the pull request is merged.

olsak:
I did nothing with this PR, I only accepted another PR [use local variables to aviod table lookups](https://github.com/olsak/OpTeX/commit/eae0a6fd14cc6f802e218c2622acacbc6a174dcd) as you can see in my master branch.

Udi-Fogiel:
might be a web page cache problem since I pushed a fix only recently. Can you try remove the browser cache and reload the page? 
This is what I'm seeing:
image

I did see the conflicts, but resolved them. If still things don't work I'll revert the latest commit and try to merge master again.

PR #214, opened 2025-11-28
hisyntax-lua: fix multiline comment

Udi-Fogiel:


olsak:
The "long brackets" are `[[` or `[=[`, or `[====[` etc. They are various "levels" of long brackets. The PR now implements only level zero. On the other hand, we can read from `hisyntax-lua.opm`:
```
% missing: [[, [=[, [==[ strings, " ' in string of other kind
```
so, multiline strings are not implemented at all. Maybe, we can implement at least level 0 and 1 of these strings:
```
\_replfromto {[[}{]]}         {\x S{[[#1]]}}%       strings [[...]]
\_replfromto {[=[}{]=]}       {\x S{[=[#1]=]}}%     strings [=[...]=]
```
It is probably not necessary to solve multiline comments too much, because they are intended for hiding parts of code during debugging process, no for presentation of code in the printed format.

Thank you, I'll apply this PR in this state.

PR #220, opened 2026-02-15
improve trick 170

Udi-Fogiel:
This commit shows how to define more general
\ul command using lua-ul, which can be used
to underline, heighlight and overline.

In addition it adds it to optex-tricks.opm
so it could be loaded on the fly.

An example:
```tex
\transparency=1
{\ul Text}
{\ul[height=1.75ex,depth=.75ex,color=\Yellow,] Text}
{\ul[height=.55ex, depth=-.45ex,color=\Red,flags=over] Text}
```
image

Note that control the color of the rule one can also do
```tex
{\Red\ul{\Black Text}}
```

So the color and transparency keys are not really needed, but I though it is more convenient.
Do you think they are redundant?

The `\transparency=1` is in the example because of a bug in `\transparency` with values<1.

### DIFF
--- a/optex/base/optex-tricks.opm
+++ b/optex/base/optex-tricks.opm
@@ -1323,11 +1323,55 @@ end,'global') % we want to keep the meanig after group
 
 \_nspublic \ismatch ;
 
+\_trick 0170 \ul \uldefaults ;
+%%%%%%%%%%%%
+
+\_directlua{
+  local funcs = lua.get_functions_table()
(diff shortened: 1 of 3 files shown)


Udi-Fogiel:
or maybe adding an item?

olsak:
Yes, we need to reformulate all relevant items. May be: better is to include all these cases to the first item. I leave it to your decide.:)

PR #237, opened 2026-05-26
fix set_node_color bug

Udi-Fogiel:
* \_colorcnt should be incremented globally (as done from the TeX side with \_incr)

* Normal category codes should be used to set the color macros (to avoid problems with spaces)

### DIFF
--- a/optex/base/optex.lua
+++ b/optex/base/optex.lua
@@ -13,6 +13,7 @@
 -- \enditems
 
 local fmt = string.format
+local string_upper = string.upper
 
 local node_id = node.id
 local node_subtype = node.subtype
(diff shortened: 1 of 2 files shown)


Udi-Fogiel:
Not really related to this PR, but you can avoid the dependency of the file `fontawesome5-mapping.def` in OpTeX trick 12 by reading the glyphs names directly from the font data, something like
```
\initunifonts
\font\tenfafree      =[FontAwesome5Free-Solid-900]
\font\tenfabran      =[FontAwesome5Brands-Regular-400]
\font\tenfafreeregul =[FontAwesome5Free-Regular-400]
\protected\def\faregul#1{{\let\tenfafree=\tenfafreeregul #1}}
\protected\def\fafree{\tenfafree\resizethefont}
\protected\def\fabran{\tenfabran\resizethefont}
\beglua
for name, value in pairs(font.getfont(font.id("tenfafree")).resources.unicodes) do
    if ('-alt' ~= string.sub(name, -4)) and value >= 256 and name:match("^[a-z-]+$") then
          tex.sprint("\\protected\\def"
                .. string.gsub('\\fa-' .. name, '-(%w)', string.upper)
                .. "{{\\fafree\\char" .. value .. "}}")
    end
end
for name, value in pairs(font.getfont(font.id("tenfabran")).resources.unicodes) do
    if ('-alt' ~= string.sub(name, -4)) and value >= 256 and name:match("^[a-z-]+$") then
          tex.sprint("\\protected\\def"
                .. string.gsub('\\fa-' .. name, '-(%w)', string.upper)
                .. "{{\\fabran\\char" .. value .. "}}")
    end
end
\endlua

\faHouseUser, \faAngry, \faApple, \faStopwatch, \faWifi

\bye
```
This way you only need the font files.

olsak:
Thank you for the tip with Awesome5 fonts.

PR #240, opened 2026-06-14
remove sentinel ^^J from \_prepareverbdata

Udi-Fogiel:
To my understanding this is only a sentinel for `\_printverb`, to avoid and extra line, so I don't think it should be part of the data that is processed for highlighting or other processes.

### DIFF
--- a/optex/base/lang-data.opm
+++ b/optex/base/lang-data.opm
@@ -1,6 +1,6 @@
 %% This is part of the OpTeX project, see http://petr.olsak.net/optex
 
-\_codedecl \_langdata {Language dependent data <2026-05-17>} % only en, cs preloaded in format
+\_codedecl \_langdata {Language dependent data <2026-07-20>} % only en, cs preloaded in format
 
 \_langdata en {English} % -----------------------------------------------
 \_langw en  Chapter      Table       Figure       Subject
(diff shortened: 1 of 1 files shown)


olsak:
Sorry, I don't understand why this commit is usable. It removes final `^^J` before `\_prepareverbdata` and adds it back at the end of the `\_prepareverbdata` output when it is used for printing. If we have

    \_def\_prepareverbdata#1#2{\_def#1{#2}}

then there is no difference. But if someone changes this macro then the difference may occur. And I am not sure if someone (who redefines it) rely on the fact that there is this sentinel `^^J`. If it is true then the macro will be broken after this commit.

Udi-Fogiel:
> Sorry, I don't understand why this commit is usable. [...]

> But if someone changes this macro then the difference may occur. 

Yes, this is the point :) so I guess you get the point.

> And I am not sure if someone (who redefines it) rely on the fact that there is this sentinel `^^J`. If it is true then the macro will be broken after this commit.

The doc says
> The `\_startverb` puts the scanned data to `\_prepareverbdata`. It sets the data to `\_tmpb` without
changes by default, but you should re-define it in order to do special changes if you want. (For example,
\hisyntax redefines this macro.)

I would argue that the data in `\_tmpb` is changed because of that extra ^^J. This ^^J is not part of the user input, and requires a special treatment, which is a bit annoying.

But backward compatibility is a consideration (although personally I would prefer to simplify things when possible).

Udi-Fogiel:
> But backward compatibility is a consideration (although personally I would prefer to simplify things when possible).

And if I redefine `\_prepareverbdata` I could always strip the extra `^^J` myself there and add it back in the end after finishing, so it is not a big deal I guess.

You can close this PR if you prefer to keep things as they are.

olsak:
Maybe your suggestion gives sense, it is more clear. But I am not able to remember why I added `^^J` before `\_prepareverbdata`. Maybe here is a hidden reason I don't see it now. I have to study the code of `\hisyntax`, `\algol`, doc, `\_printverb` etc. in order to more understand why the `^^J` is present here. This is not simple for me. I need some time...

Udi-Fogiel:
Thanks. There is absolutely no rush

PR #243, opened 2026-07-19
add OpTeX trick 0187 (sorting with lua-uca)

Udi-Fogiel:


olsak:
I did copy-paste the code between `<pre>` and `</pre>` into uca-sort.tex fle and when I \input uca-sort then I get the error:
/uca-sort.tex[\directlua]:1: 'end' expected near 'key'. Unfortunately, the message doesn't report the error position exactly, so I didn't do any more experiments.

olsak:
My mistake, I replaced `&lt;` by `<` and the code works. Thank you.

